PR #78, opened 2015-11-05
added support for VIM (syntax highlighting etc)

gamato:


gamato:
I should have said "brand new support" :-)

Issue #182, opened 2015-11-25
add support for the unary + operator


gamato:
This fails during parsing — number n=+1.0n;

tethal:
```
unhandled QORE System exception thrown in TID 1 at 2015-11-26 07:34:33.138111 Thu +01:00 (CET) at <stdin>:1
PARSE-EXCEPTION: syntax error, unexpected '+', expecting variable reference or namespace-scoped global variable reference
```

Seems that qore does not support unary + operator.

Issue #31, opened 2015-10-05
Empty statement not supported


tethal:
Qore does not allow empty statements:

    if (1) 
    	;
    else {
        ...
    }
    
    while (busy_wait())
    	;

Introducing empty statements will probably result in ambiguities due to `{}` having two meanings (empty block vs empty hash):

    if (x) {};

Could be interpreted as either:

    if (x) hash();		# statement has no effect

Or two statements (with both doing nothnig):

    if (x) {
    }

    ;

(Tested with version 0.8.12, git hash: 2ee5c954b5c88e6e56f0435b9ad75dcb5c324e73)

Issue #30, opened 2015-10-05
Inconsistent meaning of empty curly braces


tethal:
Empty statement block is not allowed in the top level scope:

    if (1) {
        {}			# OK
        {};         # error, statement has no effect
    }
    
    {};             # error, statement has no effect
    {}		        # error, unexpected $end

(Tested with version 0.8.12, git hash: 2ee5c954b5c88e6e56f0435b9ad75dcb5c324e73)

PR #244, opened 2015-12-10
Resolved reduce/reduce conflits, support for unary plus and empty stmt

tethal:
refs #30 
refs #31 
refs #182


tethal:
@pavelkveton please have a look at this as well

tethal:
Conflicts were resolved by making sure that whenever a statement is expected, a curly opening brace always means the beginning of a block, it is never a hash. It makes some obscure constructs illegal:

```
{a: sub(){}}.a();
```

Issue #305, opened 2015-12-22
add strict mode for parse_float() and parse_number()


gamato:
What if Qore::parse_float() and Qore::parse_number() would accept 3rd arg "bool strict = False" that would turn on strict parsing that might a) return NOTHING or b) throw an exception (both in cases when input looked e.g. "1x2") ?

Atm they both silently drop non-numeric part of the input, which is surely not wanted in some / many cases.

Issue #164, opened 2015-11-23
allow for leading / trailing decimal point


gamato:
Float type does not allow leading or trailing decimal point. Why?

Consider allowing for leading / trailing decimal point (as is common in other languages).

PR #315, opened 2015-12-28
Add is_int, is_float, is_number utility functions

gamato:
These are strict matching, meaning is_float() won't match int literals.


tethal:
I think that this deserves some discussion as it is not clear how are `int`, `float` and `number` defined. For example:
- are leading/trailing spaces really part of a number? 
- should `-0` be allowed for an integer?
- what about int values that cannot fit into 64 bits?
- also, there are some special float values such as `NaN` and `+/-Inf`, shouldn't we consider them as acceptable floats?

Minor style issues (though these are probably not documented yet):
- there should be no spaces between function/method name and opening parenthesis
- `%strict-args` should be used

pavelkveton:
yes, there should be a discussion about it. I'd say we should stick to 
some "standard" definition of what is / is_not a number - C functions 
strtod/strtol may be a good start...
Pavel

gamato:
These are utility functions to tell if a string conforms to a numeric.
Whitespace is most usually (and safely) ignored.
Minus and plus are most usually ignored in front of zero, that is -0 == 0 == +0
(Even some kind of affinity, which is not a case here, would not make it non-numeric.)
is_int() does not deal with size of argument, it simply says it represents a number.
We could have is_int32() etc if we ever need them, but this is not the case here / yet.
I'm not sure if special float values have an accepted representation and they are ignored here.
If Qore does have support for such special cases and if we ever need them, we can implement them.

tethal:
The problem here is "most usually". We can discuss indefinitely what each of us considers an int or float. IMO there needs to be some correlation between `is_int(x)`, `int(x)`, `parse_int(x)`, what Qore parser accepts and whatever else is out there. Looking at the use of `is_int()` in #317, the intention seems to be something like "`is_int(x)` is `True` iff `int(x)` does not fail and produces a correct result" but then of course we have a problem with the definition of `int(x)`. For a value >= 2<sup>63</sup> the Csv parser will probably not do what one would expect.
I think that this and #300, #305, #316 and even #164 need to be considered together in a consistent way. For example there would be no need to have these functions if #305 was implemented since CvsUtil could use the strict version of `parse_int()` instead.

gamato:
By most usually I meant how it's done elsewhere.  Contrary to Qore other languages usually fail on int("123y") but they are perfectly fine with int(" 123 ").  Btw, parser does not care how much whitespace there is around a numeric literal, too.
I don't think a size restriction should be included here.  One can do `if (is_int(n) and n<2**32)` (or `n<100` or whatnot).  Also, feeding a too big value to int() may fail and that is perfectly okay.
Btw, there's no size limit for integer in CSV spec (if there's such thing).  One can try int() and fall back to number(), if necessary.  But being able to distinguish between integer and float is useful.

As for the linked tickets:  #164 is Qore parser issue, #305 are Qore functions (and not available yet), #300 and #316 are more of business issues.  They are related but not necessarily that close.  I can imagine scenarios where simple casting, e.g. int(n), would not tell me if the input is broken or just too big.  Also, right now there's no way to really distinguish different numerics.  So from my point of view all of the tickets complement each other.

pavelkveton:
well, "other languages" "usually" say that also locale affects what is a 
number or not. I'd rather stick to some standard (like ISO or so) than 
to "other languages" - or, let's chose one language and implement is 
EXACTLY as it is done there.

anyway, if there is is_int() function and int() convertor from string to 
int, then I'd expect as a developer that int() should fail exactly when 
is_int() returns false. The same with is_float()/float() and maybe 
others. Put like this, we could make it more effective and do something like

int to_int(string s, \bool is_int)

where is_int is optional "return value" - if used, than it would get 
true/false depending on ability to convert 's' to integer. if not used, 
to_int could throw an exception on conversion failure.

is_int could then be an (almost) alias to this function. In either case, 
the string would get parsed/analyzed only once, compared to 
constructions like

if (is_int(something))
{
     int x = int(something);
     // do with 'x'
}

anyway, I don't know any ISO or other standard for numeric formats, 
would have to dig it out...

Pavel

gamato:
OK, I'd like to make sure this is moving somewhere, so here's the list with concerns and proposals what to do about them.
1. leading/trailing spaces
   White space is generally ignored, when converting a variable, in a language parser, on a command prompt, etc. I've just tested 5 languages I've got "laying around" and all of them ignore whitespace -- Oracle, Postgres, Python, Qore, Shell.  Note that Postgres requires an explicit convert mask and yet it ignores unspecified whitespace.
2. should -0 be allowed for an integer
   Should +0 be allowed?  From my experience -0 == 0 == +0 and again I tested this with all the 5 languages and all of them behaved the same way.
3. int values that cannot fit into 64 bits
   The purpose of my functions is not to test against any size / precision constraints.  That's best to be done separately.  Also, I can imagine a use cases where I might want to classify input without actually converting it to a specific type / representation.  We could add e.g. is_int32() but we don't need it now and it's easy to do: `n < exp2(32)`. I think it's better to have small simple functions that can be combined to build something bigger.
4. special float values such as NaN and +/-Inf
   We don't need them here.  These are special values defined by floating-point standard and produced during certain mathematical (FPU coprocessor) operations.If we ever need them in this context, which I doubt, they could be added without API change.
5. no need to have these functions if #305 exists
   Partially true, but wouldn't cover all use cases.  parse_int() wouldn't tell us if an input is broken or just too big.  And if latter, user could for instance use arbitrary precision library.
6. also locale affects what is a number or not
   I'm not aware of Qore supporting this.  Other languages often have special means for such cases.  If we ever need it, we can add it.  But we don't right now and it's not a valid show stopper.

Perhaps it's all just because you might have assumed these functions purpose is different than I intended.  If we need functions like you described, we can add them later, even w/o changing API.  Perhaps those could / should be part of base math library.  But I think there's a place for generic utility functions I created and they do exactly what I meant them to -- they tell me if a string is a valid representation of a numeric value or not.  Nothing more and nothing less.  Thanks and regards.

tethal:
Thanks for the explanation.

From my POV the whole problem is hidden in the first sentence of your last paragraph: 

> you might have assumed these functions purpose is different than I intended

My problem is that you are extending the public API of a module and it is not clear from the documentation of the functions what they are supposed to do and thus I had to make assumptions.
I understand that

> they do exactly what I meant them to 

but I still don't see this as a good definition.

Now to the points:
1. ok, no problem, but I'd prefer this to be documented
2. ok, this was in fact my confusion, since -0 != +0 in some contexts, but there is no problem in the context of parsing an integer
3. well this one is interesting:
   
   ```
   $ qore -n -X '"65536" < exp2(16)'
   True
   ```
   
   so there's obviously something more going on. Anyway my point here was because of the use in CsvUtil where the input `s = "412348761274836712983468712461346"` would pass the `is_int(s)` test, but would be evaluated to `9223372036854775807` by `int(s)`, making the purpose of these functions even more confusing.
4. ok, but FYI Qore has these special values (case insensitive) `@NaN@` and `@Inf@` for floats and even `@NaN@n` and `@Inf@n` for numbers
5. again, this comment was based on the use in CsvUtil, where you first check the input and then parse it, whereas parsing it directly and throwing an exception when the parser fails makes more sense to me. Also, having only one implementation reduces the number of bugs, documentation and inconsistencies. But this whole point is probably the least important one.
6. this is just another example that `what is a number` is a difficult question, but I agree that we probably don't want to go into discussion about locales
7. (added by me) every integer is a number so I would expect `is_int(x) => is_number(x)`. Otherwise `is_number()` should be named `is_number_but_no_an_int()` or it should be clearly documented, but I would still consider this a pitfall. For floats it is not so obvious but in [Qore](https://qore.org/manual/current/lang/html/data_type_declarations.html) we do consider `int` a subtype of `float` and `float` a subtype of `number`. The expression `float("34")` evaluates to 34.0 so I'd say that `is_float("34") == True`

To sum up, the issue here is documentation and I'd still consider points 3 and 7 as open.

Thanks.

gamato:
3) This looks like a Qore issue.

$ qore -X 'exp2(16).type()'
"number"

$ qore -X 'number("65536") < exp2(16)'
False

$ qore -X '"65536" < exp2(16)'
True

Anyway, I needed something for CsvUtil bug fix, but when I thought about it I decided to implement these general purpose functions.  Implementation in CsvUtil may change but I believe these functions are generally useful to have.

Then again, if we need implementation specific behaviour, #305 would likely do, but it's not enough for other use cases as I explained before.

4) Yes, those values are important for numeric floating point calculations, but I have yet to see them used in other contexts.  And as I said, their support can be added if ever needed.

7) Right, I thought about that, too. But if is_number() would match floats and integers too, there wouldn't be a way to distinguish them. And so now we can either combine them as I did in CsvUtil or use parse_number() / number() to consider subtypes too.

Right, documentation can and likely should be improved, I agree.

tethal:
I already created an issue #330 for relational operators which seem broken on `number` types.

omusil24:
I just discovered that the string type's class (`<string>`) also has a function `bool <string>::intp()` which checks if the string can be converted to integer:
https://qore.org/manual/current/lang/html/class_qore_1_1zzz8stringzzz9.html#ace7f5b6ac2e7b9f4886c9e153edd46d7
I didn't look at the code so I don't know what the function considers as an integer.

gamato:
Thanks, I haven't noticed either.  But it works the same way as int() does -- "123qaz".intp() returns True.

tethal:
Thanks for the documentation update.

3) Now that the documentation clearly states that the check is not done, the discussion is just about the usefulness of `is_int()`, which is clearly subjective and since I don't feel the need for this function in the first place, I can't provide any more input to this. So unless the others have something to say, I guess we can keep this.

7) I still consider it misleading when a function named `is_float()` 'rejects' values such as `4` or `-42`. These **are** floats. I don't get your argument

> But if is_number() would match floats and integers too, there wouldn't be a way to distinguish them

Wouldn't these be equivalent?

```
is_number_that_rejects_floats_and_ints(x) iff is_number_that_accepts_floats_and_ints(x) && !is_float_that_accepts_ints(x)
is_float_that_rejects_ints(x) iff is_float_that_accepts_ints(x) && !is_int(x)
```

Also, a pattern like this would still work:

```
if (is_int(x)) {
    #do something
} else if (is_float(x)) {
    #do something
} else if (is_number(x)) {
    #do something
} else {
    #do something
}
```

And they would be actually easier to use in #317.

gamato:
OK.
Would adding an argument, e.g. 'bool strict = False', resolve it?
Or a different naming, e.g. is_pure_float() or is_strict_float() or is_just_float() or whatnot ?

tethal:
Well, if the question is "Would these be better names for a function that rejects integers?" then yes and my preference would be is_pure_float().

gamato:
7) Are they really floats? Can you provide your definition of float please? Because from my experience any such numeric literal without attached type is by default considered integer datatype and dealt with accordingly. That means it is stored and handled differently than "real" floating point numeric values. This is common in most computer language interpreters / compilers I've worked with.

8) I'm not sure with this one yet but I'd prefer extra argument that would select which kind of behaviour we want from the functions. Later we could even add one more argument that might make sure the literal could be stored as Qore value (as was suggested previously).

omusil24 (changes requested):
Please remove the spaces between function names and parentheses. In all of Qore code and tests, it's standard to not have any spaces there.

pavelkveton:
no problem on my side - and being a prerequisite for the other pull request, I agree.

omusil24 (approved):

PR #317, opened 2015-12-28
Feature/csvutil strict numerics

gamato:


tethal:
depends on #315

omusil24 (approved):

pavelkveton:
I'd more like have the exception thrown directly from int(), float() or whatever other method. But overall idea is ok, I agree.

PR #319, opened 2015-12-29
Empty hash in the hash map operator

tethal:
Described in #236.

The actual implementation is in the third commit, the others are just code style.


tethal:
Please see my comment in #236

PR #331, opened 2015-12-29
Add ordinal() utility function

gamato:
Returns string with partially textual representation of ordinal integer value, e.g. ordinal(21) returns "21st".


davidnich:
looks good

PR #342, opened 2015-12-31
Add plural() utility function

gamato:
Returns string with number and proper singular/plural form of noun.

### DIFF
--- a/examples/test/qlib/Util/plural.qtest
+++ b/examples/test/qlib/Util/plural.qtest
@@ -1,0 +1,40 @@
+#!/usr/bin/env qore
+
+%requires ../../../../qlib/Util.qm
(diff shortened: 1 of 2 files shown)


pvanek:
Hmm, it is english only solution FYI.

In cs_CZ there is eg:
1 pivo 
2 piva
5 piv
21 pivo
...

pvanek:
It should be implemented lile this:
http://doc.qt.io/qt-5/i18n-plural-rules.html

gamato:
Actually, there are languages with more than 3 rules, IIRC Mozilla has it implemented in Gecko.

Anyway, we can add additional parameter(s) and modify the function to support more than English, but right now we do not have, nor support, multilingual strings / messages in Qore code (which was the main motivation for this function).

PR #532, opened 2016-02-18
Feature/sqlutil commit api

pvanek:


pvanek:
@davidnich @gamato  - first draft

Issue #552, opened 2016-02-22
qit build break when compiling qore on ubuntu 14.04


pavelkveton:
I'm opening it against qore and not the build-tools since finally it appears that the problems are in qore. Here is the patch required to succesfully compile:

diff --git a/cmake/unix-config.h.cmake b/cmake/unix-config.h.cmake
index d5b45bd..1be6383 100644
--- a/cmake/unix-config.h.cmake
+++ b/cmake/unix-config.h.cmake
@@ -112,7 +112,7 @@
 #cmakedefine HAVE_GETHOSTBYNAME_R
 #cmakedefine HAVE_STRTOIMAX
 #cmakedefine HAVE_STRERROR_R
-#cmakedefine STRERROR_R_CHAR_P
+#cmakedefine STRERROR_R_CHAR_P 1

 /\* OpenSSL */

diff --git a/include/qore/intern/config.h b/include/qore/intern/config.h
index b6344cc..c449fc4 100644
--- a/include/qore/intern/config.h
+++ b/include/qore/intern/config.h
@@ -38,7 +38,7 @@
 #endif

 #ifdef HAVE_UNIX_CONFIG_H
-#include "unix-config.h"
+#include <qore/intern/unix-config.h>
 #else
 #if defined(_WIN32) && !defined(__CYGWIN__)
 #include "win32-config.h"
diff --git a/include/qore/intern/qore_socket_private.h b/include/qore/intern/qore_socket_private.h
index 3d3fc2e..294fb04 100644
--- a/include/qore/intern/qore_socket_private.h
+++ b/include/qore/intern/qore_socket_private.h
@@ -46,9 +46,9 @@
 #include <openssl/ssl.h>
 #include <openssl/err.h>

-#if defined HAVE_POLL_H && defined HAVE_POLL
+#if defined HAVE_POLL
 #include <poll.h>
-#elif defined HAVE_SYS_SELECT_H && defined HAVE_SELECT
+#elif defined HAVE_SELECT
 #include <sys/select.h>
 #elif (defined _WIN32 || defined __WIN32__) && ! defined **CYGWIN**
 #define HAVE_SELECT 1

pvanek:
a candidate to cmake branch

pavelkveton:
could be a candidate on feature/cmake-build, but for now it is a 
showstopper for me in the development...

niclasr:
[...]

PR #554, opened 2016-02-23
refs #552: fixes in cmake and file inclusion to compile on ubuntu 14.04

pavelkveton:
using qit on qore-less Ubuntu 14.04 fails to compile qore due to missing dependencies and incorrect preprocessor defines. The fix possibly could break other builds, though...


tethal:
I'm not sure about this - @davidnich can you have a look please? Some of these are your recent changes. Thanks

PR #557, opened 2016-02-23
CsvUtil: writers have encoding support

pvanek:
A pretty critical stuff for rossini/hyperion


omusil24:
Hotfixes are for the old branch, not for develop... Use bugfix branch.

PR #737, opened 2016-04-12
Fixed configure so it can find dlfcn-win32 library.

omusil24:


omusil24:
Merged since the failing build was caused by unrelated Mysql problem.

PR #854, opened 2016-05-12
Bugfix/852 is_writable, is_readable

omusil24:
Bugfix for issue #852. Used open(2) for checking as has been discussed before.

@gamato @pvanek @davidnich please check.


omusil24:
Util.qm module is using is_writable to check if tmp directory exists. It should be using is_dir instead. That leads to a bit of a problem. There's no way to distinguish if a directory is writable/readable/executable or not.

davidnich:
there is already a solution for this for windows in the c++ code - a temporary file is created in the directory and then deleted - what about that solution?

omusil24:
That only tells you if you can write in that directory. It doesn't tell you if you can read it or "execute" it. Also, it shouldn't be in is_writable or is_readable, since according to documentation, these functions are for files only, not directories.

> bool Qore::is_readable    (   string  path    )  
> Bug:
> on Windows this function always returns True for directories; this will be fixed in a later version of Qore

This should also be addressed.

davidnich:
@omusil24 these functions are intended also for directories (as you can see when all the tests are now failing because is_writeable() is failing for /tmp), so please treat any documentation to the contrary as a documentation bug

gamato:
Patch fine by me.

Re tmp_location() -- The question is what should the function tell us.  If it's just "Returns a path for temporary files", then perhaps no checks are necessary.  If it should give us a functional temporary directory, then perhaps it's best to create a file to make sure it's doable (and maybe even read it back).  Otherwise there's a risk the path isn't usable (traversable) or files cannot be created etc.

PR #902, opened 2016-05-25
Improve output of comparison of numeric values

gamato:
For numeric values, when really verbose or difference not visible, compute and show the difference.


davidnich:
I greatly prefer %y to %n, because the output with complex data structures is much more compact and very easy to read - is there any reason for using %n here?

gamato:
Please note this is for numeric types only. Is there a (significant) advantage of %y in this case?

davidnich:
for numeric types the output is identical, but I believe that %y should be used anyway as a best practice - in my opinion we should deprecate %n for %y and not use it in Qore modules

PR #938, opened 2016-06-01
Bugfix/915 numerical comparisons

tethal:


tethal:
Majority of the code has been reviewed, but I'm not sure about the change in QoreTreeNode

PR #1058, opened 2016-07-13
Added StreamReader + edited InputStreamLineIterator to use it internally.

omusil24:


include/qore/StreamReader.h
```diff
@@ -0,0 +1,363 @@
...
+      return i;
+   }
+
+   DLLLOCAL virtual const char* getName() const { return "StreamReader"; }
+
+private:
+   DLLLOCAL int64 fillBuffer(qore_size_t bytes, ExceptionSink* xsink) {
+      assert(bufSize + bytes <= bufCapacity);
+      int64 rc = in->read(buf + bufSize, bytes, xsink);
+      bufSize += rc;
```
tethal, line 291: it is not safe to use the value of rc if an exception is raised by in->read()

include/qore/StreamReader.h
```diff
@@ -0,0 +1,363 @@
...
+      return i;
+   }
+
+   DLLLOCAL int64 readi2(ExceptionSink* xsink) {
+      short i = 0;
+      if (!prepareEnoughData(2, xsink))
+         return 0;
+      i = reinterpret_cast<short*>(buf)[0];
+      shiftBuffer(2);
+      i = ntohs(i);
```
tethal, line 180: htons would make more sense here

include/qore/StreamReader.h
```diff
@@ -0,0 +1,363 @@
...
+      bufSize(0),
+      buf(0) {
+      is->ref();
+      if (!enc->isAsciiCompat()) {
+         xsink->raiseException("WRONG-ENCODING-ERROR", "StreamReader does not support ASCII-incompatible encodings");
+         return;
+      }
+
+      // +1 is added to the real capacity for terminating null-character.
+      buf = new char[bufCapacity + 1];
```
tethal, line 63: memory leak

include/qore/StreamReader.h
omusil24, line 180: ntohs makes more sense because it means that the source data is in big-endian, independently of the end result.

tethal, line 180: ok, you're right sorry

PR #1104, opened 2016-07-25
Feature/869 file streams - std i/o support, mode

tethal:


tethal:
@davidnich I'm not sure about the naming convention for global constants (stdin_stream or stdinStream?)

PR #1125, opened 2016-08-02
CsvUtil backward compatibility broken for single-row-type format (#1124)

tmandys:
for backward compatability when only single type is defined then type is always resolved, i.e. trailing optional fields are assigned as NOTHING and excess fields are ignored


omusil24:
@tmandys Did you rebase your branch? What happened that there are 300 commits in this pull request?

pavelkveton:
this request should be opened against 0.8.12 branch.

Issue #1134, opened 2016-08-03
qore.pc generated with wrong libdir


pavelkveton:
qore.pc.in contains reference to @libdir@ which is set by configure to ${exec_prefix}/lib. The libqore library cannot be usually found on this path, since it is installed to libdir as defined in Makefile.am, which is ${exec_prefix}/lib${LIBSUFFIX} (which is platform dependent).

as such, if pkgconfig takes place when configuring for qorus (or whatever tool using pkgconfig to look for qore), it won't find the library where needed (for example, qorus uses hard-coded $OMQ_DIR/lib... ON TOP OF pkgconfig).

PR #1135, opened 2016-08-03
refs #1134 added a hack that allows for setting proper path in qore.p…

pavelkveton:
…c; note that the entire situation is unfortunate, since qore.pc is generated from qore.pc.in either by cmake OR by autotools, whatever is actually used, and each of the build systems use slightly different way to initialize the variables to be substituted

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1751,6 +1751,13 @@ else
 fi
 AC_SUBST([myprefix])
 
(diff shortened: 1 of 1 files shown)


pavelkveton:
please don't delete the branch, we need to merge to develop

pavelkveton:
bad pull request, need to merge to 0.8.11.1 first

PR #1138, opened 2016-08-03
Bugfix/1134 qore pc wrong libdir

pavelkveton:


pavelkveton:
another incorrect request, need to merge to 0.8.12 first

PR #1474, opened 2016-12-05
SchemaReverse: implementation

pvanek:
as requested - pure qore repo branch


omusil24:
The test does not work with PostgreSQL. I believe this is not a wanted behaviour since the module should be DB-agnostic.
@pvanek Can you fix it?

Issue #1428, opened 2016-11-21
SqlUtil: schema alignment: default values can be complex statement instead of simple value only


pvanek:
for example table containing an expression to calculate default value - perfectly valid in Oracle:

```
create table foobar (
  id number not null,
  org_id number default TO_NUMBER ( DECODE ( SUBSTRB ( USERENV ( 'CLIENT_INFO') ,1,1 ) , ' ',NULL,SUBSTRB ( USERENV ( 'CLIENT_INFO') ,1,10 ) ) )
);
```

There is no way how to implement it in schema hash now

PR #1492, opened 2016-12-09
Feature/1428 sqlutil default values

pvanek:
this patch allows us to use native functions in default values like eg.:

```
create table t (
   dt date default to_date('1.1.2000', 'dd.mm.yyyy') not null
);
```

in different DBMS dialects.

It also reduces copypasted option-for-drivers mergers

Refs: #1428


davidnich (approved):
looks good

gamato (approved):
Hi. It looks good except maybe some typos, but I would like to see one more test (for e.g. Postgres) to make sure it works properly. Thanks.

PR #1499, opened 2016-12-12
Changed math.h includes to cmath in 0.8.12 because of Windows build

omusil24:


omusil24:
There will probably be conflicts when merging to develop... I'll handle that.

PR #1666, opened 2017-02-14
1660 SqlUtil op_in() for 0.8.12

gamato:
This is a "rebase" of #1661 for branch 0.8.12


pvanek (changes requested):
"union all" cannot be used here. Let's assume if there will be a huge list with repeating elements (1,2,3, ..., 1,2,3, 1,2,3). Union all will cause duplicated rows returned. We have to use plain "union" which will cause additional performance damage of course...

How it will work with orderby and / or aggregate functions?

Also I suggest to extend docs to say that using huge in lists is a sign of potentially bad design of an application.

gamato:
I'm not sure how to "directly" reply to change request from the review above, so here's my regular comment:

I thought about the same thing.  Eventually I lean towards processing duplicates instead of performance impacts of union.  Duplicates can be, and often are, removed in user code.  And even if they are not, it seems better option than sorting for unions.

The documentation was extended with notes as you suggest.  Maybe the wording can be better?  Please suggest or update the branch, I'm fine with it. Thanks.

pvanek (approved):

PR #1876, opened 2017-05-22
Merged astparser module to Qore

omusil24:
Added the module to `qore/modules/astparser` directory. When building Qore with Cmake, the module is automatically built. When building with autotools, the module can be manually built with its own cmakefile.


omusil24:
Only those last few commits need to be checked because there are changes related to integration with Qore. The rest are all QLS commits.

davidnich (approved):
for me it's fine - when Petr approves, I'll merge

pvanek (approved):

PR #1879, opened 2017-05-23
astparser is buildable when there is no qore installed yet

pvanek:


omusil24:
@pvanek Why did you remove the C++11 check, "CMAKE_BUILD_TYPE" code and the `project()` and library/project version code?

pvanek:
all these props are inherited from upper directory. project() etc is suitable only of the module is buildable standalone, which is not true anymore. This module must be always built with qore.

PR #1888, opened 2017-05-25
refs #1866 added missing quoting to the comment string

pavelkveton:


pavelkveton:
incorrect branch name

PR #1891, opened 2017-05-25
Bugfix/1886 adding column comment with single quote when creating new schema in postgresql leads to an exception

pavelkveton:


davidnich (approved):
Looks good to me

pavelkveton:
assigning to myself and going to merge since it was already approved by David anyway

PR #1935, opened 2017-06-07
Feature/cmake fixes

omusil24:
@pvanek


pvanek:
**critical:**

include/qore/intern/unix-config.h:
```
#define HASH_NAMESPACE
```

is not correct probably. The define probably requires some value. When I change it (manually) to 
```
#define HASH_NAMESPACE std
```
then it's compilable. See src code: include/qore/hash_map_include.h for reference

Error for reference:
```
[ 25%] Building CXX object CMakeFiles/libqore.dir/QC_AbstractSmartLock.cpp.o
In file included from /export/home/pvanek/src/qore/qore/include/qore/intern/ConstantList.h:205:0,
                 from /export/home/pvanek/src/qore/qore/include/qore/intern/QoreClassIntern.h:37,
                 from /export/home/pvanek/src/qore/qore/include/qore/intern/QoreObjectIntern.h:54,
                 from /export/home/pvanek/src/qore/qore/include/qore/intern/QoreClosureNode.h:36,
                 from /export/home/pvanek/src/qore/qore/include/qore/intern/QoreLibIntern.h:407,
                 from /export/home/pvanek/src/qore/qore/include/qore/Qore.h:225,
                 from /export/home/pvanek/src/qore/qore/build/QC_AbstractSmartLock.cpp:32:
/export/home/pvanek/src/qore/qore/build/include/qore/hash_map_include.h:6:32: error: ‘unordered_map’ in namespace ‘::’ does not name a type
 #define HASH_MAP HASH_NAMESPACE::unordered_map
                                ^
/export/home/pvanek/src/qore/qore/include/qore/intern/ConstantList.h:208:9: note: in expansion of macro ‘HASH_MAP’
 typedef HASH_MAP<const char*, ConstantEntry*, qore_hash_str, eqstr> cnemap_t;
```


**minor:**

- Q: make install does not have docs in dependencies. Do we want something like "install-docs"? Or do we expect that docs are handled (rpm, deb) on the packager level?

- "make dist" macro exists, but it's not used in the CMakeLists.txt

- "make uninstall" is missing for AST
```
CMake Warning at cmake/QoreMacros.cmake:187 (message):
  Uninstall script: no file:
  /export/home/pvanek/src/qore/qore/modules/astparser/cmake/cmake_uninstall.cmake.in
Call Stack (most recent call first):
  modules/astparser/CMakeLists.txt:64 (qore_binary_module)


CMake Warning at cmake/QoreMacros.cmake:208 (message):
  file does not exist:
Call Stack (most recent call first):
  modules/astparser/CMakeLists.txt:64 (qore_binary_module)
```
I suggest to do it as in case of Doxyfile.in, the file should be shared for all cmake modules


- Also maybe to remove this warning
```
CMake Warning at CMakeLists.txt:783 (message):
  Please do not use CMake for production/real builds.  CMake support is
  experimental and it does not contain all required setup.
```

- what about to delete all automake files? ;)

davidnich:
> what about to delete all automake files? ;)
when it’s 100% finished and we are absolutely sure we don’t need it anymore then yes

however we have solaris builds (which weren’t working before with cmake) windows cross-compiles - rpms etc still todo

let’s not rush on this please

cau
david


--
David Nichols
Qore Technologies, s.r.o.
Přemyslovská 1939/28 130 00 Prague - 3 Žižkov
Czech Republic
mobile: +420 777 279 144
https://www.qoretechnologies.com

PR #1955, opened 2017-06-16
Bugfix/50 escaping docs

omusil24:


omusil24:
This has to wait until #1949 is merged to develop.

PR #1960, opened 2017-06-19
Added method to AstTree to 'export' the tree structure to Qore hashes and lists

omusil24:


pvanek (changes requested):
good start, but it still needs some improvements.

a structure returned from getNodesInfo() 

   - what a "nodetype" is? Are there some enums available?
   - "loc" cannot be returned like this. It requires additional parsing to get proper values. Some real int values (4 attributes) should be used I think

can we make AstParser methods static? To be used like this?

AstTree t = AstParser::parseString(CODE);

instead of

AstParser p();
AstTree t = p.parseString(CODE);

BTW what is a class AstTree for? It just holds "printer/output" methods. So if these AstParser::parse return a list/hash, we can live without it...


Should not we remove AstTree::printTree? It's somehow a helper/unused method for real life, right?

omusil24:
`nodetype` is the type of node, i.e. `decl`, `expr`, `stmt`, `po`, `name`. In C++ there is enum obviously, but Qore doesn't have that...

Locations... Ok, no problem. Any preferred format/hash keys? (start.line, start.col, end.line... or firstLine, firstCol, lastLine...)

AstParser methods could be static theoretically but the reason that it is a class is that it's also a log/logger class storing the errors which happened during parsing (which you should check after each parse). For this, there are two methods - `getErrorCount` and `getDiagnostics` (returns diagnostics in the format suitable for LSP).

AstTree class is there to represent the tree in Qore (store the internal ASTTree C++ class) + provide export and printing functions and anything that might be suitable for it (and which wouldn't make sense in AstParser for example).

`AstTree::printTree` is a helper/debug print method. I think it's useful for that. I don't feel like removing it yet.

pvanek:
I vote for start_column, start_line, end_column, end_line

pvanek (approved):

PR #1993, opened 2017-07-03
CMake automatic module versions

niclasr:
small fix, get the module versions from ModuleManager.h instead of setting them in CMakeLists.txt as well. This version does not use get_define.sh and works under msys.


davidnich (approved):
looks good

@niclasr please assign to @omusil24 when ready to merge

niclasr:
I think @omusil24 should try this pull request with his mingw64 setup since it replace code in #682 that is needed for his environment. I just tried get-define.sh on my msys and it worked thats why I want him to try this pull request on his mingw64 before merging.

PR #2228, opened 2017-09-25
fixes #2227: sqlutil: cop_over is broken in statements with join or s…

### LINKED ISSUE #2227: sqlutil: cop_over is broken in statements with join or superquery. partitionby and orderby clauses are not expanded to fully qualified column names
caused by #2203

pvanek:
…uperquery. partitionby and orderby clauses are not expanded to fully qualified column names


omusil24:
Release notes are missing, no?

pvanek:
no. It's a fix for feature introduced last friday (it has relnote on its own)

PR #2240, opened 2017-09-26
Fixed astparser to correctly handle EOF in block comments (same as in #2176)

omusil24:


omusil24:
@davidnich Can you merge this please?

PR #2277, opened 2017-10-10
fixes #2276: sqlutil binary: use BulkSqlUtil to speed up load of huge…

### LINKED ISSUE #2276: sqlutil binary: use BulkSqlUtil to speed up load of huge tables

pvanek:
… tables


davidnich (approved):
cool, nice change

Issue #2377, opened 2017-11-04
"C:relative\path\" considered absolute


geckon:
There are relative paths on Windows that qore considers absolute:

```
$ qore -nX "Qore::absolute_path_windows('c:path\to\file.txt')"
True

$ qore -nX "Qore::absolute_path_windows('c:..\path\to\file.txt')"
True
```

Quoting [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx):

> If a file name begins with only a disk designator but not the backslash after the colon, it is interpreted as a relative path to the current directory on the drive with the specified letter.

PR #2381, opened 2017-11-05
refs #2377 Paths like "C:path\to\file.txt" are relative on Windows.

geckon:


omusil24:
Actually, the release notes are missing... Made #2382 to fix it.

PR #2557, opened 2017-12-21
QUnit more verbose output to provide info about number/float

tmandys:


tmandys:
Added testAssertionValue() overloaded number/float/auto

Issue #3008, opened 2018-09-19
Add clang-format formatting file


petrycta:
Create .clang-format file according to code standards and refactor code

PR #3009, opened 2018-09-19
Add .clang-format file #3008

petrycta:


petrycta:
I have processed required changes (includes formating, arguments in functions and do not allign inside #ifdef blocks) and prepared .clang-format file. Now we should decide when the refactoring should be done, to make it less painful

omusil24:
Fixed some of the directives to better match the suggestions in the first thread and to better match our current code. And also sorted the directives alphabetically.

omusil24 (approved):

Issue #2966, opened 2018-08-29
Separate Qore module loadings


petrycta:
### Idea: to have an ability to structure complex modules into multiple files.
**Curret status:**
_One-File-One module_
**Proposed Qore Implementation:**
    _One-File-Modules_

### Implemented as directories.
```html
MyModule/
    MyModule.qm - a file with the same name as the base directory has. It contains standard module headers as in single file module. It does or it doesn't contain any public symbols (namespaces, classes, functions...)
    Class1.qc - imported automatically into QoreProgram
    functions.ql - imported automatically into QoreProgram
```

For more details see [wiki](https://bugs.qoretechnologies.com/projects/qorus/wiki/QoreModules)

Estimated time: ~5WD
Spent time: 3WD

omusil24:
Duplicate of #2562

petrycta:
closed as duplicate

PR #3033, opened 2018-10-11
Separate qore module loading #2966

petrycta:


lib/ModuleManager.cpp
```diff
@@ -846,6 +857,58 @@ void QoreModuleManager::loadModuleIntern(ExceptionSink& xsink, const char* name,
...
+    modulePath += feature;
+    modulePath += ".qm";
+
+    QoreProgram* p = pgm ? pgm : path_pgm;
+    if (!p) {
+        p = mpgm;
+        if (!p)
+            p = getProgram();
+    }
+    const char* td = p ? p->parseGetScriptDir() : 0;
```
sejvlond: use nullptr here instead of 0

examples/test/qore/requires/MyModule/SimpleModuleClass.qc
```diff
@@ -0,0 +1,18 @@
+# -*- mode: qore; indent-tabs-mode: nil -*-
+# Qorus SimpleModuleClass class definition
+
+%new-style
```
sejvlond, line 4: use 
```
%new-style
%enable-all-warnings
%require-types
%strict-args
```

examples/test/qore/requires/MyModule/SimpleModuleFunctions.ql
```diff
@@ -0,0 +1,11 @@
+# -*- mode: qore; indent-tabs-mode: nil -*-
+#! @file SimpleModuleFunctions.ql consists functions definitions
+
+%new-style
```
sejvlond, line 4: use 
```
%new-style
%enable-all-warnings
%require-types
%strict-args
```

examples/test/qore/requires/requiresModul.qtest
```diff
@@ -0,0 +1,39 @@
...
+        set_return_value(main());
+    }
+
+    testIncludeWholeModule() {
+        Program p(PO_NEW_STYLE|PO_REQUIRE_TYPES);
+        p.parse("
+#!/usr/bin/env /home/tpetr/devel_release/bin/qore
+# -*- mode: qore; indent-tabs-mode: nil -*-
+
+# TODO: change to some relative path
```
sejvlond: remove the TODO pls

examples/test/qore/requires/requiresModul.qtest
```diff
@@ -0,0 +1,39 @@
...
+# examples path is for Jenkins; '.' is for manual runs
+%append-module-path examples/test/qore/requires:.
+%requires MyModule
+
+MyModule::mfunc(\"\");
+", "");
```
sejvlond: fyi multi line string can also be written as

```p.parse(
    "line1 - no plus, just end quotes\n"
    "line2 - automaticly concatenated, no indentation will be there\n"
    "you have to use \n for newlines"
```
it's just fyi, here is probably beter to use your style even it's weirdly indented

lib/ModuleManager.cpp
```diff
@@ -740,6 +742,16 @@ void QoreModuleManager::loadModuleIntern(ExceptionSink& xsink, const char* name,
...
         // see if it's a user or binary module
         size_t len = strlen(name);
+        QoreString modulePath(name);
+
+        QoreProgram* p = pgm ? pgm : (load_opt & QMLO_REINJECT ? (mpgm ? mpgm : nullptr) : pholder.release());
+        if (!p) {
+            p = getProgram();
+        }
+
+        if (p)
```
sejvlond: use `{}` around the one line if

lib/ModuleManager.cpp
```diff
@@ -752,6 +764,12 @@ void QoreModuleManager::loadModuleIntern(ExceptionSink& xsink, const char* name,
 
             mi = loadBinaryModuleFromPath(xsink, name, 0, pgm, reexport);
         }
+        else if (QoreDir::folder_exists(modulePath, xsink)) {
+            qore_offset_t i = modulePath.rfind(QORE_DIR_SEP);
+            // "feature" means pure module name (e.g. "Mime", "CsvUtil" etc.)
+            const char * feature = QoreString(modulePath).extract(++i, &xsink)->c_str();
```
sejvlond: you create a new QoreString, then you callExtract, which returns a pointer to the extracted string, then you take c_str which is a const char * into the internal representation of QoreString, and then you use it. But I would say, that the temporary copy of QoreString, that is extracted should be already destructed, or not? 
@davidnich can you help? I stil don't fully understand how this works, what is returned from .extract (it's newly allocated string) and how long would it live?

doxygen/lang/120_modules.dox.tmpl
```diff
@@ -19,6 +19,16 @@
 %requires ../../directory/Module.qm
     @endcode \n
 
+    Module implementation could be separated to multiple files and stored as a directory named with the same name as the module. To require them use @ref requires "%requires ModuleName" without ".qm" extension. All files inside the directory will be automatically included at program parse time.
```
sejvlond: `All files inside ` - only some, based on the extension

doxygen/lang/125_include_files.dox.tmpl
```diff
@@ -11,5 +11,7 @@
 
     After this, any variable, function, namespace, constant, or object declared in the above file can be used by the %Qore program.
 
+    If program (SomeClass.qc or SomeFuntion.ql) is a part of separated module (see @ref qore_modules) it will be included automatically.
```
sejvlond: does it really called program?

sejvlond (changes requested):

lib/ModuleManager.cpp
davidnich: @sejvlond @petrycta yes this is a memory leak - `QoreString::extract()` returns a new object that needs to be deleted

also please run valgrind on your tests to make sure that you haven't introduced any memory leaks

sejvlond: so if extract returns a new object, shouldn't it be const and so we can call it on `modulePath` directly? or does it also change the object on which it's called?
and should it be freed with `delete` or just `deref`? thx

davidnich: > so if extract returns a new object, shouldn't it be const 

yes it should, but that is not related to this branch or issue

> and so we can call it on modulePath directly? 

it would be best

> or does it also change the object on which it's called?

it should not, it is most likely a `const` error

> and should it be freed with delete or just deref?

with `delete` as it is not a reference-counted class - `QoreStringNode` is a different story

sejvlond: > `QoreStringNode` is a different story

ah, I see
thx

examples/test/qore/requires/requiresModul.qtest
petrycta: ok, thanks, please check my new interpretation

doxygen/lang/120_modules.dox.tmpl
petrycta: changed to:
All files inside the directory will be automatically included at program parse time.

doxygen/lang/125_include_files.dox.tmpl
petrycta: changed to If such a program code (i.e. SomeClass.qc or SomeFuntion.ql)

examples/test/qore/requires/requiresModul.qtest
petrycta:
```
p.parse(
    "#!/usr/bin/env /home/tpetr/devel_release/bin/qore\n"
    "# -*- mode: qore; indent-tabs-mode: nil -*-\n"

    "# examples path is for Jenkins; '.' is for manual runs\n"
    "%append-module-path examples/test/qore/requires:.\n"
    "%requires MyModule\n"

    "MyModule::mfunc(\"\");\n"
, "");

assertEq(
    "MyModule::mfunc::EXAMPLE_C::SomeClass::cfunc\n"
    "MyModule::mfunc::EXAMPLE_C::func\n"
    "MyModule::mfunc::EXAMPLE_F::func\n"
, p.callFunctionArgs("mfunc", ""));
```

lib/ModuleManager.cpp
petrycta: resolved as:
```
std::shared_ptr<QoreString> feature(QoreString(modulePath).extract(++i, &xsink));
```
also new memory leak found here:
```
QoreListNode* fileList = moduleDir.list(&xsink, S_IFREG, &regexClassesFunc);
```
resolved as:
```
fileList->deref(&xsink);
```
now everything works cleanly, see attached screenshot with valgrind output
valgrind_confirm

petrycta:
Thanks. All review comments have been processed

lib/ModuleManager.cpp
```diff
@@ -767,8 +768,8 @@ void QoreModuleManager::loadModuleIntern(ExceptionSink& xsink, const char* name,
         else if (QoreDir::folder_exists(modulePath, xsink)) {
             qore_offset_t i = modulePath.rfind(QORE_DIR_SEP);
             // "feature" means pure module name (e.g. "Mime", "CsvUtil" etc.)
-            const char * feature = QoreString(modulePath).extract(++i, &xsink)->c_str();
-            mi = loadSeparatedModule(xsink, modulePath, feature, pgm, reexport, pholder.release(), load_opt & QMLO_REINJECT ? mpgm : 0, load_opt);
+            std::shared_ptr<QoreString> feature(QoreString(modulePath).extract(++i, &xsink));
```
sejvlond: there is no point of creating shared_ptr, you are the exclusive owner of the value (you are not passing it anywhere where the ownership has to be passed/shared), you can use unique_ptr, it's cheaper to create and it makes more sense.

lib/ModuleManager.cpp
```diff
@@ -904,6 +908,7 @@ QoreAbstractModule* QoreModuleManager::loadSeparatedModule(ExceptionSink& xsink,
         std::string fileCode = QoreDir::get_file_content(filePath);
         userModule->getProgram()->parsePending(fileCode.c_str(), feature, &xsink, &xsink);
     }
+    fileList->deref(&xsink);
```
sejvlond: instead of using deref manually, use
`ReferenceHolder<QoreListNode> fileList =  ....`

@davidnich is there some easy know how, what pointers should be deleted, what should be wrapped into reference holders? it's so hard to keep track of it, you have to go in implementation of everything... maybe some info/intro from you could help?

sejvlond (changes requested):

lib/ModuleManager.cpp
davidnich: @sejvlond if the value has a protected destructor, then it must be dereferenced, otherwise it should be deleted normally

sejvlond: and btw why isn't the value already returned as `ReferenceHolder<T>` - you would either have to accept it as a reference holder - creating or copying reference holders should be really cheap, or you would call `.release()` by yourself, but from the very declaration of the function it would be clear, that it is reference counted object, right?

lib/ModuleManager.cpp
petrycta: done

lib/ModuleManager.cpp
petrycta: done

sejvlond (approved):

PR #3076, opened 2018-10-30
Feature/3069 msys win builds

pvanek:
this set of patches makes:

- qore (and modules) compilable on Windows using MSYS2/MINGW
- qore easier to use with modules on Windows - it does not use full paths for module paths resolving.

Points to discuss/check:

- [x] does it cross compiles in David's release toolchain?
- [x] should we use it for OSX bundles too? (if we will provide a dmg for qore)

Also here is a HOWTO for Windows:
https://github.com/qorelanguage/qore/wiki/MS-Windows


davidnich (approved):

omusil24 (approved):

lib/ModuleManager.cpp
```diff
@@ -466,22 +466,80 @@ void QoreModuleManager::init(bool se) {
...
+    do {
+        buff.resize(buff.size() + MAX_PATH);
+        copied = GetModuleFileName(0, &buff.at(0), buff.size());
+    } while (copied >= buff.size());
+
+    buff.resize(copied);
+    std::string prefix_path(buff.data(), buff.size());
+    size_t found_pos = prefix_path.rfind('\\');
+
+    return prefix_path.substr(0, found_pos).append("/../").append(path).c_str();
```
sejvlond: after `substr.append.append` there is a temporary string variable (on stack), that will be destroyed immediately after exiting the context, then you take a pointer into its member `c_str` and you return it - after return, the temp string is destroyed and the pointer points to undefined memory - normally this should ends with deq fault (maybe sometimes with debug build it can work)

sejvlond: I would return the string, it's safer
also it's a pity that API of `UniqueDirectoryList` hash push_back variant only for `DLLLOCAL bool push_back(const char* str)` even that inner implementation uses `std::set<std::string>` so the char* is the copied anyway.. Maybe introduce new API `push_back(const std::string &str)` adn `push_back(std::string &&str)`, he? :)

lib/ModuleManager.cpp
```diff
@@ -466,22 +466,80 @@ void QoreModuleManager::init(bool se) {
...
+    // Let's expand it until it fits
+    std::vector<char> buff;
+    DWORD copied = 0;
+    do {
+        buff.resize(buff.size() + MAX_PATH);
+        copied = GetModuleFileName(0, &buff.at(0), buff.size());
+    } while (copied >= buff.size());
+
+    buff.resize(copied);
+    std::string prefix_path(buff.data(), buff.size());
```
sejvlond, line 511: whát? this looks really really ugly, you are resizing vector of chars and continuously trying to getModuleFileName over and over? where is the `GetModuleFileName` I cannot find it, cannot it be updated to match better?

sejvlond (changes requested):

lib/ModuleManager.cpp
pvanek: It's a Windows system call. 
You should remember that MAX_PATH is often not enough, especially on modern systems with long pathnames.
https://stackoverflow.com/a/33613252

lib/ModuleManager.cpp
pvanek: no idea what to do here. Should I copy the char*? Or to return a std::string and call c_str() in the `moduleDirList.push_back()` calls?

sejvlond: if you would copy char* then the caller will have to free it, second variant is okay, return string and call .c_str when passing it to push_back (it will be copied there), or :) create the new API for moduleDirList that would accept `constr string &` and `string &&` (the last variant will create no copy)

lib/ModuleManager.cpp
sejvlond: ah, let me look then

lib/ModuleManager.cpp
pvanek: check it now, please

lib/ModuleManager.cpp
sejvlond: 1, you can use string, instead of vector, (no additional copy, no need for resizing at the end etc)
```
    std::string prefix_path;
    DWORD copied = 0;
    do {
        prefix_path.resize(prefix_path.size() + MAX_PATH);
        copied = GetModuleFileName(0, &prefix_path.at(0), prefix_path.size());
    } while (copied >= buff.size());
```

2, this is a win api, that uses `LPSTR`, `DWORD` and we are passing `char *`, `size_t` and then we are comparing `dword` with `size_t`, doesn't it raises some warnings? I don't know if it's okay (maybe some issues with multibyte chars? I don't know how lpstr handles them)

3,
> If the function fails, the return value is 0 (zero). 

maybe we should check for it?

sejvlond:
please fix the conflicts, also there are commits that shouldn't be there, probably brought by some merges?

lib/ModuleManager.cpp
sejvlond: 1, what do you tihnk?
2, ok, proably we can omit it, should be more or less safe
3, I would add a check that would raise an exception?

omusil24:
Fixed the conflicts. Also the new commits are there because other PRs were merged to develop in the meantime. That's fine.

PR #3097, opened 2018-11-09
Fix #3096: double % in parse directives docs

### LINKED ISSUE #3096: Formatting error in parse directives docs
`%allow-injection` and `%allow-weak-references` are listed with `%%` in the docs which is probably a copy and paste error.
https://docs.qore.org/current/lang/html/parse_directives.html#allow-injection

geckon:


davidnich (approved):
thank you, @geckon

Issue #3107, opened 2018-11-13
Improve docs regarding hash keys


geckon:
The [docs on hashes](https://docs.qore.org/current/lang/html/container_data_types.html#hash) says that keys are strings which isn't entirely true...
```
$ qore -ne 'hash h = {"aaa": "b", 3: 4}; printf("%N\n", h);'
hash: (2 members)
  aaa : "b"
  3 : 4

$ qore -ne 'hash h = {"aaa": "b", 3: 4}; printf("%N\n", h{3});'
4

$ qore -ne 'hash h = {"aaa": "b", 3: 4}; printf("%N\n", h.3);'
4
```

This can be confusing at least. I understand that there probably is some kind of conversion done implicitly but it confuses me that when I print the whole hash, it "knows" that I used an integer as a key to define that value so it's probably not just a one-way conversion.

geckon:
It is in fact a one-way conversion. Only strings are used inside and those integer key above are displayed by %N.

```
$ qore -ne 'hash h = {"aaa": "b", 3: 4}; printf("%N\n", h."3");'
4

$ qore -ne 'hash h = {"aaa": "b", "3": 4}; printf("%N\n", h."3");'
4

$ qore -ne 'hash h = {"aaa": "b", "3": 4}; printf("%N\n", h.3);'
4
```

martin-zemek:
Anything in a position of a hash key is converted to string. You can have `qore -X '{{"a":1}: 3}'`  which gives `{'': 3}` since `{"a":1}.toString() = ''`. or even `qore -X '{NOTHING: 3}'` gives `{'': 3}` since `NOTHING.toString() = ''`.
On the other hand,  `{[1,2]: 3}` produces an exception since `[1,2].toString()` produces an exception.

Actually, the docs on hashes does not literally say that keys must be provided as strings. The formulation "Hashes are containers that associate values to a string key ..." allows for the interpretation:  "keys act as strings".

(By the way, the fact that `NOTHING` can be a hash key makes it possible to chain access to undefined hash indexes, so not only `{"a": 1}.b` is possible but also `{"a":1}.b.c.d` (yes, you'll get a warning but also a result).)

martin-zemek:
As to: ' ...   it "knows" that I used an integer ...':
It does not know that you used an integer, it just prints the string "3" as 3.

geckon:
Thanks for more comments @martin-zemek.

[...]

PR #3128, opened 2018-11-20
refs #3107: Specify hash keys data types better.

geckon:


davidnich (approved):

martin-zemek:
The new sentence is:
If anything else than a string is used as a key, the value is one-way converted to a string and the string
value is used as a key. This can sometimes have unexpected results and therefore only string keys are recommended.

I would write only:
If anything else than a string is used as a key it is converted to a string and the string is used as a key.

geckon:
@martin-zemek Is it because you think that the results can not be unexpected or because the docs should be strictly descriptive without giving recommendations?

I mean.. I do agree that it should be possible to use anything hashable as a key but it should be used as it is, not converted to a string. Also anything not hashable (like hashes or lists) should not be possible to use as a key.

Things like these may be predictable if you know how it works but it's far from being intuitive...
```
$ qore -ne 'hash h = {"1": "xxx"}; h{1} = "aaa"; printf("%N\n", h);'
hash: (1 member)
  1 : "aaa"
```
```
qore -ne 'hash h = {NOTHING: "xxx"}; h{("a": "b")} = "aaa"; printf("%N\n", h);'
warning encountered at <command-line>:1
INVALID-OPERATION: key number 0 (starting from 0) in the hash is type 'nothing', which cannot be converted to a string, therefore will always evaluate to an empty string at runtime
next warning:
warning encountered at <command-line>:1
INVALID-OPERATION: the right side of the expression with the '.' or '{}' operator is hash<string, string>, which cannot be converted to a string, therefore will always evaluate to an empty string at runtime
hash: (1 member)
   : "aaa"
```

I know there is a warning  in the second case but this should not even be possible IMO.

Until we support other data types properly, I think we should only recommend strings. Just my 2 cents...

martin-zemek:
I only meant that once it is exactly documented it is no longer unexpected (in other words user gets unexpected result only if he has not read the documentation). But you can leave it there, it's no harm, I only said that I wouldn't write it.

omusil24 (approved):

martin-zemek (approved):

Issue #3120, opened 2018-11-16
FsUtil: Add overwrite argument to copy_path() and copy_tree()


geckon:
After discussion with @pvanek I think it makes sense to add an `overwrite` argument to `copy_path()` and `copy_tree()` functions in FsUtil module similarly to what `copy_file()` has.

geckon:
`copy_tree()` with overwrite will remove the destination tree and the copy the source. If anyone doesn't like this solution or find it confusing, please let me know.

geckon:
This will actually affect the current behavior as well. Right now when the destination directory exists, `copy_tree()` will create a subdirectory where the copy will be created - e.g. when `/some/path` exists (and is a directory), `copy_tree("/path/to/source", "/some/path");` will create a directory `/some/path/source` and there the copy will be created. Otherwise it will copy the contents of `/path/to/source` to `/some/path`. This is sort of the behavior of Unix `cp`.

We can change that so that the source path always will be the exact path of the tree root. If it exists (and overwrite is True), it will be removed and then copied. With overwrite set to False it will fail.

Is that an acceptable solution? Let's discuss along with #3123.

PR #3148, opened 2018-11-27
Implement #3120: Add overwrite flag to copy functions (FsUtil)

geckon:


davidnich (approved):
your changes look fine.  it's not a part of this pull request, but the code takes advantage of a bug in Qore that needs to be fixed where Qore ignores missing type declarations in catch blocks - specifically I mean `catch (err)` - this should be `catch (hash<ExceptionInfo> err)` ideally

geckon:
Thanks @davidnich, I filed #3149.  

Is the bug in Qore (ignoring missing type declarations) reported or should I create an issue for that one as well? I couldn't find it but maybe you know about an existing issue?

sejvlond (approved):

pvanek (approved):

PR #3150, opened 2018-11-27
refs #3121: Implement copy_dir_structure()

geckon:


davidnich (approved):

qlib/FsUtil.qm
```diff
@@ -808,4 +808,54 @@ public namespace FsUtil {
...
+        *hash<StatInfo> src_stat = hlstat(source);
+        if (!exists src_stat) {
+            throw "DIR-STAT-ERROR", sprintf("Couldn't stat source path '%s': %s", source, strerror());
+        }
+
+        # create destination directory
+        if (path_exists(destination)) {
+            throw "PATH-EXISTS-ERROR", sprintf("Destination path '%s' already exists.", destination);
+        }
+        mkdir(destination, 0700);
```
omusil24, line 842: There should be a mask parameter. Not hardcoded 700.

geckon: I discussed this with @omusil24. It can be hardcoded because it's later set to match the mode of source.

qlib/FsUtil.qm
```diff
@@ -808,4 +808,54 @@ public namespace FsUtil {
...
+                # ignore symlinks
+                if (is_link(src_path)) {
+                    continue;
+                }
+
+                copy_dir_structure(src_path, join_paths(destination, name), next_depth);
+            }
+        }
+
+        chmod(destination, src_stat.mode);
```
sejvlond, line 857: this may not be called if somewhere inside the recursion there will be an error, so part of the structure will be copied and it's permissions won't be changed, right?

is that a problem? should we add the `chmod` into `on_exit` or should we just state this in documentation, or should we ignore it?

geckon: That's true. On the other hand if there is an exception, the caller doesn't know what was copied and what wasn't and so the state of `destination` is undefined. Therefore I think this doesn't really matter, does it?

I can add a comment stating something like "if the call fails, the structure is only partially copied and its state is undefined" but it seems kind of clear to me.

Another option would be to remove whatever was copied on any error so either the structure is copied completely or not at all. 

BTW, something similar is happening in `copy_tree()` with `fail_immediately` set to `True` but there the last mentioned option would be problematic because it also has an `overwrite` option (and so removing `destination` on error would not necessarily recover the previous state.

sejvlond: ok, just wanted to talked about that, this is fine for me

geckon: ok, thx

Issue #3343, opened 2019-03-08
Zeyos Rest connection missing required option error


turlyalz:
```
HTTP: REST DBG: class ["user", "post", "User exception: /usr/share/qore-modules/ZeyosRestClient.qm:259 (Qore): ZEYOSRESTCONNECTION-ERROR: missing required option \"http_version\" in option argument for connection \"zeyostest\" with url \"zeyosrest://cloud.zeyos.com/alzhan/\" (zeyostest)
```

PR #3345, opened 2019-03-08
refs #3343: fix use only ZeyosOptions not all Options in the connecti…

turlyalz:
…on constructor


davidnich (changes requested):
it would be great to have a test for the issue

PR #3380, opened 2019-04-23
Fixed #3378 sorting strings with different lengths

### LINKED ISSUE #3378: sort() doesn't work correctly for strings of different length
This isn't right:

```
$ qore -nX 'sort(("hc", "h"))'
list: (2 elements)
  [0]="hc"
  [1]="h"
$ qore -nX 'sort(("h", "hc"))'
list: (2 elements)
  [0]="h"
  [1]="hc"
```

Maybe the different length isn't the defining property that connects these cases together but it seems to work properly if the strings have the same length.

omusil24:


davidnich (changes requested):
@omusil24 can you please add tests for the comparison operators that were also fixed with this fix - i.e. 
```
assertTrue("ab" > "a");
assertTrue("a" < "ab");
```
etc

omusil24:
@davidnich added the tests.

davidnich (approved):

PR #3874, opened 2020-04-02
Feature/3872 map auto method

turlyalz:


davidnich (changes requested):
in fact the crec arg should not be present in `mapAuto()`, because it's ignored in some cases.  To be consistent, the argument should be omitted entirely IMO.

also please add a release note in `doxygen/lang/900_release_notes.dox.tmpl` for this change

davidnich (approved):

PR #4115, opened 2020-12-09
Feature/4106 alpine build

martin-zemek:


davidnich (changes requested):
the `HAVE_PTHREAD_GET_STACKSIZE_NP` is used in `qore_thread_intern.h`

won't this change mean that it's not used on platforms that need it with an autotools build?  For example this is used on Darwin to get the stack size

martin-zemek:
> the `HAVE_PTHREAD_GET_STACKSIZE_NP` is used in `qore_thread_intern.h`
> 
> won't this change mean that it's not used on platforms that need it with an autotools build? For example this is used on Darwin to get the stack size

the `HAVE_PTHREAD_GET_STACKSIZE_NP` has not been removed, it's only been moved higher to keep the alphabetical order

PR #5151, opened 2026-01-14
Add EventNotifier class for efficient cross-thread signaling

### LINKED ISSUE #5150: Add EventNotifier abstraction for efficient cross-thread signaling
## Summary

Add an `EventNotifier` class that provides an efficient, platform-optimal mechanism for cross-thread event notification, to be used with `EventLoop`.

## Background

Currently, `AsyncSocketIoController` uses a pipe for signaling between threads. While this works, it has some drawbacks:
- Requires two file descriptors (read/write ends)
- Less efficient than platform-specific mechanisms
- Potential for race conditions between pipe read/write and command processing

## Proposed Solution

Create an `EventNotifier` class with platform-optimal implementations:

### Linux
Use `eventfd()`:
- Single file descriptor
- Counter-based semantics (multiple notifications coalesce)
- Designed specifically for event notification
- More efficient kernel implementation

### BSD/macOS  
Use `kqueue` with `EVFILT_USER`:
- Native kqueue user events
- No file descriptor overhead
- Integrated with the event loop

### Fallback
Use pipe for platforms without specialized support.

## API Design

```cpp
// C++ implementation
class EventNotifier {
public:
    EventNotifier(ExceptionSink* xsink);
    ~EventNotifier();
    
    int fd() const;           // fd to register with EventLoop (Linux/fallback)
    void notify();            // signal the notifier (thread-safe)
    void acknowledge();       // consume notification after poll returns
    bool isValid() const;
};
```

```qore
# Qore binding
class EventNotifier {
    constructor();
    notify();                 # Signal the notifier
    acknowledge();            # Consume notification
}

# Integration with EventLoop
EventLoop::add(EventNotifier notifier);
```

## Benefits

1. **Performance**: Uses most efficient mechanism per platform
2. **Simplicity**: Single abstraction instead of managing pipes
3. **Reliability**: Reduces race condition surface area
4. **Resource efficiency**: Single fd on Linux vs two for pipes

## Implementation Plan

1. Add `EventNotifier` C++ class with platform detection
2. Add Qore bindings in `QC_EventNotifier.qpp`
3. Integrate with `EventLoop` class
4. Update `AsyncSocketIoController` to use `EventNotifier`
5. Add tests

## References

- nginx uses `eventfd` on Linux: `src/event/modules/ngx_epoll_module.c`
- Linux eventfd(2) man page
- kqueue EVFILT_USER documentation

davidnich:
## Summary
- Adds `EventNotifier` class for efficient cross-platform event notification
- Uses `eventfd` on Linux (efficient counter-based signaling) and pipes on macOS/BSD
- Integrates with `EventLoop` for poll-based waiting
- Updates `AsyncSocketIoController` to use `EventNotifier` instead of pipe-based signaling

## Changes
- New `QoreEventNotifier` C++ class with platform-specific implementations
- New `EventNotifier` Qore class exposed via QPP bindings
- `EventLoop` extended with `add(EventNotifier)` and `del(EventNotifier)` methods
- `AsyncSocketIoController` simplified to use `EventNotifier`
- Unit tests for `EventNotifier`

## Test plan
- [x] EventNotifier unit tests pass (3 test cases, 13 assertions)
- [x] AsyncSocketIo tests pass (14 test cases, 41 assertions)
- [x] Valgrind shows no memory leaks
- [ ] CI passes on all platforms

Closes #5150

🤖 Generated with [Claude Code](https://claude.ai/code)


davidnich:
Regarding the Copilot suggestion to change `return QoreValue();` to `return;`:

This suggestion is **incorrect** for QPP files. The QPP preprocessor generates C++ wrapper functions that return `QoreValue`, so `return QoreValue();` is the correct pattern. The `nothing` return type in the Qore declaration means the method returns no value to Qore code, but the underlying C++ wrapper still needs to return a `QoreValue` (an empty one in this case).

The existing code in the same file (lines 178, 206, 234) uses the same `return QoreValue();` pattern.

PR #5157, opened 2026-01-17
feat: improve OpenAI responses streaming and context

davidnich:


davidnich:
Addressed Copilot comments that were actionable. Remaining notes: (1) buildPath leading-slash normalization is intentional to ensure exactly one leading slash; removing the regex would preserve multiple slashes. (2) Chunked SSE parsing already clears the buffer because StringOutputStream::getData() releases the internal buffer, matching ServerSentEventClient behavior, so no explicit clear is needed.

davidnich:
Processed new Copilot comments: clarified context metadata precedence, simplified buildPath, fixed doc section IDs, added override guidance for conversation store, and added error field in stream events. The 'stream': True example is correct Qore boolean syntax (capitalized). Chunked SSE buffer clearing is handled by StringOutputStream::getData() (it resets the buffer), matching ServerSentEventClient behavior.

PR #5163, opened 2026-01-23
Add mailgun dataprovider

nickmazurenko:


davidnich (changes requested):
add an entry in `doxygen/lang/120_modules.dox.tnpl` and a note in `doxygen/lang/900_release_notes.dox.tmpl`

davidnich (approved):

PR #5171, opened 2026-01-28
add support for oauth2 implicit grant type in restclient

### LINKED ISSUE #197: Add Implicit Oauth Grant for Trello integration to work
# Add OAuth2 Implicit Grant Support for Trello and Similar Services

[Qorus Gitlab Branch](https://git.qoretechnologies.com/qorus/qorus/-/tree/feature/197-implicit-grant?ref_type=heads)

## Problem

The current OAuth2 implementation only supports the `authorization_code` grant type, which requires a code exchange step. Some services like Trello use OAuth2 implicit grant (or OAuth 1.0a-style simple auth) where the access token is returned directly in the URL fragment after user approval.

**The core challenge:** URL fragments (`#access_token=xxx`) are never sent to servers by HTTP specification. When Trello redirects back with `https://server/oauth2/redirect#token=xxx`, the server only receives `/oauth2/redirect` without the token.

Additionally, Trello has non-standard OAuth behavior:
- Uses `key` parameter instead of `client_id`
- Uses `return_url` instead of `redirect_uri`
- Does not pass through the `state` parameter

### 1. Fragment Capture via HTML Page

When the OAuth2 redirect handler receives a request with `state` but no `code`, it returns an HTML page with JavaScript that:
1. Reads `window.location.hash` to capture the token from the URL fragment
2. POSTs the token back to the server
3. Server processes the token and completes the OAuth flow

### 2. Implicit Grant Type Support

- Added `implicit` as a valid `oauth2_grant_type`
- Created `OAuth2ImplicitAuthRequestOptions` - implicit grant doesn't require `oauth2_token_url` or `oauth2_client_secret`
- Updated validation to use appropriate required options based on grant type

### 3. Provider-Specific Workarounds

- Embed `state` in `redirect_uri` for providers that don't pass it through
- Add `return_url` parameter for Trello compatibility
- Template variable substitution in `oauth2_auth_args` (e.g., `key: '{{oauth2_client_id}}'`)

### 4. Cloud Credentials Support

- Trello app no longer has hardcoded API key
- `oauth2_client_id` comes from cloud configuration (like Slack)
- Users can optionally override with their own key

## Files Changed

| Repository | File | Changes |
|------------|------|---------|
| qore | `qlib/RestClient.qm` | Added implicit grant support, template substitution for auth args |
| qorus | `Classes/QorusRestApiHandlerPublic.qc` | Fragment capture HTML page + POST handler |
| qorus | `Classes/QorusRestApiHandlerV7.qc` | Implicit grant validation |
| module-v8 | `qlib/TypeScriptActionInterface/TypeScriptAppRestConnection.qc` | Skip token URL validation for implicit |
| module-v8 | `ts/src/apps/trello/index.ts` | Use implicit grant, cloud credentials |
| module-v8 | `ts/src/apps/trello/constants.ts` | Remove hardcoded API key |

## How It Works

```
1. User clicks "Authorize"
2. Browser redirects to: trello.com/authorize?key=xxx&return_url=...&callback_method=fragment
3. User approves
4. Trello redirects to: server/oauth2/redirect?state=xyz#token=abc
5. Server receives only ?state=xyz (fragment not sent)
6. Server returns HTML page with JavaScript
7. JavaScript reads window.location.hash → #token=abc
8. JavaScript POSTs token to server
9. Server saves token, marks connection as authorized
10. User redirected back to UI
```

## Testing

- [x] Trello OAuth flow works end-to-end
- [x] Token captured from URL fragment
- [x] Connection marked as authorized
- [x] API calls work (ping, list boards, etc.)
- [x] Existing authorization_code apps still work (GitHub, Slack, etc.)
- [x] Cloud credentials used when available

nickmazurenko:


davidnich (changes requested):
instead of reusing `RCF_OAUTH2_AUTH_CODE` for implicit auth, we should define:
- `RCF_OAUTH2_IMPLICIT` to be set for connections requiring implicit auth
- `RCF_OAUTH2_AUTH` that will be set for both auth code and implicit auth

In Qorus then we will check only for `RCF_OAUTH2_AUTH` in a separate change

Also if it's part of the OAuth2 spec that implicit auth should have the state set in the `redirect_uri` then the implementation is fine as it is, otherwise we need to implement a new RestConnection option for this that will set for Trello

This needs a release note in the current, unreleased RestClient version (2.3.3) and in `doxygen/lang/900_release_notes.dox.tmpl`

davidnich (approved):

